PR #2, opened 2021-06-02
Make momentum a scalar quantity, and add direction vector

AndrewEdmonds11:
We want to standardize momentum as a scalar quantity in TrkAna. This PR implements the necessary changes for the MC branches (the reco branches already have momentum as a scalar quantity). All *.momx, *.momy, and *,momz branches have been replaced by *.mom, *.costh, *.phi.

Here is the list of leaf changes in the trkana tree created by TrkAnaReco.fcl:
 - momentum of `StepPointMC` at entrance, middle and exit of tracker:
     - `demcent.momx`, `demcent.momy`, `demcent.momz` -> `demcent.mom`, `demcent.costh`, `demcent.phi`
     - `demcmid.momx`, `demcmid.momy`, `demcmid.momz` -> `demcmid.mom`, `demcmid.costh`, `demcmid.phi`
     - `demcxit.momx`, `demcxit.momy`, `demcxit.momz` -> `demcxit.mom`, `demcxit.costh`, `demcxit.phi`
 - momentum of `GenParticle`:
     - `demcgen.momx`, `demcgen.momy`, `demcgen.momz` -> `demcgen.mom`, `demcgen.costh`, `demcgen.phi`
 - momentum of `PrimaryParticle`:
     - `demcpri.momx`, `demcpri.momy`, `demcpri.momz` -> `demcpri.mom`, `demcpri.costh`, `demcpri.phi`
 - initial momentum of `SimParticle` that produced track:
     - `demc.omomx`, `demc.omomy`, `demc.omomz` -> `demc.omom`, `demc.ocosth`, `demc.ophi`

### DIFF
--- a/inc/GenInfo.hh
+++ b/inc/GenInfo.hh
@@ -16,16 +16,16 @@ namespace mu2e
   struct GenInfo {
     Int_t _pdg, _gen; // true PDG code, generator code
     Float_t _time;  // time of this step
-    XYZVec _dir;   // direction at the start of this step
     Float_t _mom; // scalar momentum at the start of this step
+    Float_t _costh; // cos(theta), where theta is angle between particle's momentum and z-axis
+    Float_t _phi;   // azimuthal angle of particle's momentum vector
     XYZVec _pos;  // particle position at the start of this step
     GenInfo() { reset(); }
-    void reset() { _pdg = _gen = -1; _time = -1.0; _dir=XYZVec(); _mom = 0; _pos = XYZVec(); }
-    static std::string leafnames() { static std::string leaves; 
-      leaves = std::string("pdg/I:gen/I:t0/F:")+Geom::XYZnames("dir") + std::string(":mom/F:") + Geom::XYZnames("pos");
+    void reset() { _pdg = _gen = -1; _time = -1.0; _mom = 0; _costh = 0; _phi = 0; _pos = XYZVec(); }
+    static std::string leafnames() { static std::string leaves;
+      leaves = std::string("pdg/I:gen/I:t0/F:mom/F:costh/F:phi/F:") + Geom::XYZnames("pos");
       return leaves;
     }
-  };  
+  };
 }
 #endif
-
--- a/inc/GenInfo.hh
+++ b/inc/GenInfo.hh
@@ -16,12 +16,13 @@ namespace mu2e
   struct GenInfo {
     Int_t _pdg, _gen; // true PDG code, generator code
     Float_t _time;  // time of this step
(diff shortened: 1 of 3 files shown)


inc/GenInfo.hh
```diff
@@ -16,12 +16,13 @@ namespace mu2e
   struct GenInfo {
     Int_t _pdg, _gen; // true PDG code, generator code
     Float_t _time;  // time of this step
-    XYZVec _mom;   // momentum at the start of this step
+    XYZVec _dir;   // direction at the start of this step
```
brownd1978, line 19: making dir an XYZVec is in principle correct, assuming we can access the functions (phi(), theta(), ...) as well as the data members in root.  Have you checked if those are available, or can you provide the libraries to load so that they work?  Do you know if that will work with uproot?  If not, I suggest to replace this with cos(theta) and phi, as those are the most relevant physical variables.

inc/TrkInfo.hh
```diff
@@ -70,31 +70,32 @@ namespace mu2e
...
-    XYZVec _omom;   // origin momentum
+    XYZVec _odir; // original direction
+    Float_t _omom;   // origin momentum (scalar)
     MCRelationship _prel; // relationship if this tracks primary particle to the event primary
  
     TrkInfoMC() { reset(); }
     void reset() { _ndigi = _ndigigood = _nactive = _nhits = _nambig = _pdg = _gen  = _proc= -1;  _otime=0.0;
       _prel = MCRelationship();
-      _opos = _omom = XYZVec(); }
+      _opos = _odir = XYZVec(); _omom = -1;}
```
brownd1978, line 80: why are 2 statements on 1 line?

inc/GenInfo.hh
soleti: As far as I know there is now to access the methods from uproot (see https://github.com/scikit-hep/uproot3/issues/418). If there is a file available produced with this version I can look at it and verify what works and what doesn't.

AndrewEdmonds11: It looks like we only write out the three floats of XYZVec and not the XYZVec itself so there's no access to the methods. I will change it to cos(theta) and phi.

inc/TrkInfo.hh
AndrewEdmonds11: No reason. I will change it

AndrewEdmonds11:
I have changed from dirx, diry and dirz to costh and phi. I have edited the initial comment of this PR to match

brownd1978 (approved):

PR #28, opened 2021-08-11
Updated Genealogy Info

AndrewEdmonds11:
I went a bit further with adding genealogy information than the original PR so here is an updated message:

In this PR I have added the ability to configure the depth of the MC genealogy that is stored in the trkana tree. The new branches are:
 - `demsim` : MC information for the particle that created the track
 - `demcparent`: MC information for the parent of the particle that created the track
  - `demcgparent`: MC information for the grandparent of the particle that created the track
  - `demcggparent`: MC information for the great-grandparent of the particle that created the track
  - additional levels insert an extra `g` in the branch name
 
This PR also remove the `demcgen` branch

The depth of the genealogy is set to 5 by default for the candidate branch (`de`). MC information for the supplemental branches (`ue`, `dm`) is not on by default. The depth can be changed with the fcl parameter:
```
physics.analyzers.TrkAnaNeg.candidate.options.genealogyDepth : 5
```
where depth = 2 stores the information of the particle that created the track and its parent.

Here is an example output for a CeEndpointMix dataset:

```
root [2] trkana->Scan("demcsim.pdg:demcparent.pdg:demcgparent.pdg:demcggparent.pdg:demcgggparent.pdg")
************************************************************************
*    Row   * demcsim.p * demcparen * demcgpare * demcggpar * demcgggpa *
************************************************************************
*        0 *        11 *        13 *      -211 *      3122 *      2212 *
*        1 *        11 *        13 *      -211 *      2212 *      2212 *
*        2 *        11 *        13 *      -211 *      2212 *        -1 *
*        3 *        11 *        13 *      -211 *      2212 *        -1 *
*        4 *        11 *        13 *      -211 *      2212 *        -1 *
*        5 *        11 *        13 *      -211 *      3122 *      2212 *
*        6 *        11 *        13 *      -211 *      3122 *      3212 *
*        7 *        11 *        13 *      -211 *      2212 *      2112 *
*        8 *        11 *        13 *      -211 *      2212 *      2212 *
*        9 *        11 *        13 *      -211 *      2212 *        -1 *
*       10 *        11 *        13 *      -211 *      2212 *        -1 *
*       11 *        11 *        13 *      -211 *      2212 *        -1 *
*       12 *        11 *        13 *      -211 *      2212 *        -1 *
*       13 *        11 *        13 *      -211 *      3122 *      2212 *
```

=============================
Original PR message;

In this PR, I have added the parent and grandparent information with new branches `demcparent` and `demcgparent`. This is just a quick solution so that people can start using it.

There will be some further development work to clean up the code, and I will think about the best way to implement going back an arbitrary number of generations. This will also include checks to make sure that the genealogy hasn't been compressed away. For the time being, going back 2 generations should be safe.

### DIFF
--- a/src/InfoMCStructHelper.cc
+++ b/src/InfoMCStructHelper.cc
@@ -103,15 +103,23 @@ namespace mu2e {
     tshinfomc._doca = pca.dca();
   }
 
-  void InfoMCStructHelper::fillSimAndPriInfo(const KalSeedMC& kseedmc, const PrimaryParticle& primary, SimInfo& priinfo, SimInfo& siminfo, SimInfo& parentinfo, SimInfo& gparentinfo) {
+  void InfoMCStructHelper::fillAllSimInfos(const KalSeedMC& kseedmc, std::vector<SimInfo>& siminfos, int n_generations) {
+    auto trkprimary = kseedmc.simParticle().simParticle(_spcH)->originParticle();
+
+    for (int i_generation = 0; i_generation < n_generations; ++i_generation) {
+      fillSimInfo(trkprimary, siminfos.at(i_generation));
+      if (trkprimary.parent().isNonnull()) {
+        trkprimary = trkprimary.parent()->originParticle();
+      }
+      else {
+        break; // this particle doesn't have a parent
+      }
+    }
+  }
+
+  void InfoMCStructHelper::fillPriInfo(const KalSeedMC& kseedmc, const PrimaryParticle& primary, SimInfo& priinfo) {
     auto trkprimary = kseedmc.simParticle().simParticle(_spcH);
 
-    if (trkprimary->parent().isNonnull()) {
-      fillSimInfo(trkprimary->parent()->originParticle(), parentinfo);
-    }
-    if (trkprimary->parent()->originParticle().parent().isNonnull()) {
-      fillSimInfo(trkprimary->parent()->originParticle().parent()->originParticle(), gparentinfo);
-    }
     // go through the SimParticles of this primary, and find the one most related to the
     // downstream fit (KalSeedMC)
 
@@ -128,7 +136,6 @@ namespace mu2e {
       }
     } // redundant: FIXME!
     fillSimInfo(bestprimarysp, priinfo);
-    fillSimInfo(bestprimarysp, siminfo);
   }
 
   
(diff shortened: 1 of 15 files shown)


src/InfoMCStructHelper.cc
```diff
@@ -125,20 +131,25 @@ namespace mu2e {
     fillGenInfo(bestprimarysp, geninfo);
   }
 
+  
   void InfoMCStructHelper::fillGenInfo(const art::Ptr<SimParticle>& gp, GenInfo& geninfo) {
+    if(gp.isNonnull()){
```
brownd1978, line 136: When the genealogy is truncated at Parent, is the GP ptr null or invalid?

src/InfoMCStructHelper.cc
```diff
@@ -125,20 +131,25 @@ namespace mu2e {
...
 
-    if(gp.isNonnull()){
-      geninfo._pdg = gp->pdgId();
-      geninfo._gen = gp->creationCode();
-      geninfo._mom = gp->startMomentum().vect().mag();
-      geninfo._costh = std::cos(gp->startMomentum().vect().theta());
-      geninfo._phi = gp->startMomentum().vect().phi();
-      geninfo._pos = Geom::toXYZVec(det->toDetector(gp->startPosition()));
-      geninfo._time = gp->startGlobalTime(); 
+      geninfo._pdg = gp.pdgId();
```
brownd1978, line 145: Should we consider renaming this struct, since it is no longer related to GenParticle?   I could also imagine a single fill function taking a specific level (0=direct, 1=parent, 2=grandparent, etc) as argument, using the same struct but different object and branch.  That would make the depth of genealogy stored in TrkAna configurable, which would be a nice feature.  That's probably beyond the intent of this PR.

src/InfoMCStructHelper.cc
AndrewEdmonds11: So this `gp` should be renamed because it refers to the SimParticlePtr that is the argument of this function. To answer your question though, the truncated particle will have a valid GenParticle ptr

src/InfoMCStructHelper.cc
AndrewEdmonds11: Yes, I think we whouls rename this struct to `SimInfo` to make it clear we are looking at information from `SimParticles`. I like the idea of a single function to fill this information so we can have a configurable genealogy depth for TrkAna.

AndrewEdmonds11:
I expanded on this PR to implement everything we want for the genealogy in TrkAna. The original comment on this PR has been updated with the details

PR #33, opened 2021-10-20
Genealogy bugfix

AndrewEdmonds11:
Looking at the new CeEndpointMix MDC2020km, I found a bug relating to genealogy for events with more than one track. I saw reconstructed true DIOs (`demcsim.gen==114`) with a filled `demcgparent` branch even though these DIO events were from the background frame where we compress away the genealogy. The issue was that there was also a conversion electron reconstructed in the same event, which does have real `demcgparent` information, and so the `demcgparent` branch wasn't being reset between tracks. This PR implements a fix for this issue.


AndrewEdmonds11:
Hang on. This PR had a bunch of other changes on it that hadn't been merged yet...

PR #51, opened 2022-03-23
BestCRV-related changes

AndrewEdmonds11:
Hi Everyone,

This PR introduces some major changes in TrkAna related to the "bestcrv" hit. A list of changes is given at the end of this post but the most important is that the ***```bestcrv``` branch will now be called ```debestcrv```*** (i.e. it is now prefixed with the track type)

Before the PR is merged, the decision of which CRV hit is "best" for a given track is made inside TrkAna and uses a closest-in-time algorithm (i.e. it finds the CRV hit closest in time to the track). A future analyzer may want to use a different algorithm and so, for this reason, I have moved this algorithm out of TrkAna and into a new module called ```BestCrvHitDeltaT```. This module performs the same algorithm and creates a ```BestCrvAssns``` (which is an ```art::Assns``` between the track and the CRV hit) and writes it to the event. This ```BestCrvAssns``` data product is then read into TrkAna and used to fill the ```bestcrv``` branch. If an analyzer wants to write a new module with a different algorithm, they just need to have the module put a ```BestCrvAssns``` in the event and configure TrkAna to read it. The new module and new data product may belong in Offline, and I will discuss that with the relevant people.

There are also some new features added. It is now possible to write the "bestcrv" for your supplemental tracks (e.g. the crv hit closest in time to the upstream track), and to write the output of multiple algorithms to different branches (see the new example ```TrkAnaReco_multipleBestCrv.fcl```).

Anyway, I will leave this PR open for a little bit while I solicit comments from various people.

Thanks,
Andy

========================================

Here is a list of changes:
 - breaking changes
   - ```bestcrv``` branch is now called ```debestcrv``` in the trkana tree created by TrkAnaReco.fcl
  - new features / modules / data products
    - new ```BestCrvAssns``` data product that is an ```Assns``` between a ```KalSeed``` and a ```CrvCoincidenceCluster```
    - new ```BestCrvHitDeltaT``` module that takes the  "bestcrv" decision out of the TrkAnaTreeMaker module and creates a ```BestCrvAssns``` for the CRV hits closest in time and second-closest in time
    - can now write out the ```bestcrv``` for supplement tracks (e.g. ```uebestcrv```)
    - can now write out multiple ```bestcrv``` branches for e.g. different algorithms with user configurable branch names (see the new example TrkAnaReco_multipleBestCrv.fcl)
 - miscellaneous
   - some clean up in the prolog
   - made ```fillmc``` fcl parameter a member variable of the class module


src/BestCrvHitDeltaT_module.cc
```diff
@@ -0,0 +1,114 @@
...
+    art::InputTag _kalSeedTag;
+    art::InputTag _crvCoincidenceTag;
+  };
+
+  BestCrvHitDeltaT::BestCrvHitDeltaT(const Parameters& conf):
+    art::EDProducer(conf),
+    _kalSeedTag(conf().kalSeedTag()),
+    _crvCoincidenceTag(conf().crvCoincidenceTag())
+  {  
+    produces<BestCrvAssns>("first");
```
brownd1978, line 60: Is the intent to save the 2 'best' (smallest Dt) Assns?  In that case, why not add the DeltaT has payload to the assns, and maybe save a collection of assns, sorted to have the best first?  The # to save (or max Dt) could be a config parameter.

AndrewEdmonds11: Hi Dave. I played around with adding the Dt as payload to the Assns and making it more configurable but I couldn't get it to work. I think the module is OK as it is though. The only reason I added a second Assns was so that I could have an example for people to follow in the future (fcl/TrkAnaReco_multipleBestCrv.fcl).

PR #63, opened 2022-06-13
Add plane info

brownd1978:


oksuzian (approved):
Useful variables to suppress cosmic background.

PR #74, opened 2023-01-19
Additional variables related to Offline PR916

brownd1978:
Add leafs for new variables provided in Offline PR 916.  I also started to modernize the struct initialization, all the structs should be updated similarly.


AndrewEdmonds11 (approved):
Thanks, Dave. I will open an issue to update all the other structs so that I don't forget, and I will merge this in when Offline PR916 is merged.

PR #78, opened 2023-03-02
Small Fixes

brownd1978:
This PR removes the use of the helix() function from the main processing, as that can fail when the reco track axis is near the z axis.  It also adds some flags.  The TrkPid function needs a complete rewrite.


brownd1978:
Hi Andy, this PR improves stability running TrkAna on KinKal fits.  It also adds a few leafs in trkhitinfo that Richie needs for calibration.  It's compatible with the current Offline head.  BTW, we plan to make a new SimJob musing sometime next week.

AndrewEdmonds11 (approved):

PR #79, opened 2023-03-25
Add extra MC steps (IPA and target)

brownd1978:
I also modernized the TrkInfo structs


AndrewEdmonds11 (approved):
Thanks for this, Dave. Could you add a fcl file that implements the new feature?

PR #84, opened 2023-05-05
Update fcl to use KinKal fits

brownd1978:
This PR is a breaking change WRT MDC2020w and earlier production.


brownd1978:
We need to process this PR to be compatible with the next wave of production files which use KinKal fits.   I suggest to tag TrkAna before merging this, and tell people that they need to use that tag to work with older mcs files.

PR #98, opened 2023-09-06
Add include required by refactoring of Offline in PR1066.

kutschke:
With this change, TrkAna builds following Offline  PR1066; I don't know if other changes are needed, perhaps updates to fcl.


brownd1978:
Hi Rob, this fix is already in PR 95

kutschke:
THanks.

PR #101, opened 2023-09-12
Genealogy update

AndrewEdmonds11:
This PR updates the ```demmcsim``` branch again. Instead of a ```generation``` leaf, we now have two leaves that store ```MCRelationships```: the relationship to the track primary (the particle that created the track) is called ```trkrel```, and the relationship to the event primary (the particle that we wanted to reconstruct) is called ```prirel```. These leaves have two members ```_rel``` for the relationship (defined in Offline/MCDataProducts/inc/MCRelationship.hh, and ```_rem``` the number of generations removed the particle is.

As an example, to look at the particle that generated the track:
```
trkana->Draw("demmcsim.mom.R()", "demmcsim.trkrel._rem==0")
```
to look at its parent, you can do:
```
trkana->Draw("demmcsim.mom.R()", "demmcsim.trkrel._rem==1")
```
etc.

You can also look at track particles that weren't the intended primary with
```
trkana->Draw("demmcsim.mom.R()", "demmcsim.prirel._rel==-1")
```

### DIFF
--- a/src/InfoMCStructHelper.cc
+++ b/src/InfoMCStructHelper.cc
@@ -129,17 +129,6 @@ namespace mu2e {
     auto trkprimaryptr = kseedmc.simParticle().simParticle(_spcH);
     auto trkprimary = trkprimaryptr->originParticle();
 
-    // Add all the primary particles first
-    SimInfo sim_info;
-    for(auto const& spp : primary.primarySimParticles()){
-      // check whether we already put this primary in
-      fillSimInfo(spp, sim_info);
-      sim_info.trkrel = MCRelationship(spp, trkprimaryptr);
-      sim_info.prirel = MCRelationship(spp, spp);
-      siminfos.push_back(sim_info);
-    }
-
-
     auto current_sim_particle_ptr = trkprimaryptr;
     auto current_sim_particle = trkprimary;
     if (n_generations == -1) { // means do all generations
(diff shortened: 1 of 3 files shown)


src/InfoMCStructHelper.cc
```diff
@@ -174,6 +160,19 @@ namespace mu2e {
...
     }
+
+    // Now add all the primary particles
+    SimInfo sim_info;
+    for(auto const& spp : primary.primarySimParticles()){
+      // check whether we already put this primary in
+      fillSimInfo(spp, sim_info);
+      sim_info.trkrel = MCRelationship(spp, trkprimaryptr);
+      sim_info.prirel = MCRelationship(spp, spp);
+      if (sim_info.trkrel != MCRelationship::same) {
```
brownd1978, line 171: I think you want 'sim_info.trkrel == MCRelationship::none.  The primary could match a parent or grandparent and so already be in the siminfo vector.  A safter test would be to explicitly loop over all the siminfo entries testing !=same.

In case the particle is already in the siminfo vector, I think you want to update the prirel value.  Or does that already get filled in fillSimInfo?

brownd1978 (commented):
Thanks for re-arranging the vector with trk info first.  I had a question about how the SimInfo prirel is filled.

src/InfoMCStructHelper.cc
AndrewEdmonds11: Thanks, Dave. The prirel and trkrel values get filled in this function ```fillAllSimInfos``` instead of ```fillSimInfo``` but ```prirel``` and ```trkrel``` are set by now so I should be able to look through the SimInfos already in the vector and to check them

AndrewEdmonds11: The safer test has been included in commit [478e135](https://github.com/Mu2e/TrkAna/pull/101/commits/478e1354f95fcf713ce5644b69064665e8bab478)

PR #103, opened 2023-09-18
Fix nightly: change in CRV accessor

AndrewEdmonds11:
Nightly validation failed because ```GetAvgCounterPos()``` no longer exists. Based on the PR that introduced that change, I think the correct new accessor is ```GetAvgHitPos()``` (asking Ralf review and confirm)


ehrlich-uva:
Yes, GetAvgHitPos() is the replacement for GetAvgCounterPos(). The difference is that the updated code calculations includes the longitudinal position of the hit along the counter and not just the center positions of the counters.

ehrlich-uva (approved):

Issue #109, opened 2023-10-11
Add end position etc information to demmcsim branch


AndrewEdmonds11:
At the moment, we just have the start position information. It would be useful to have the end information too

PR #115, opened 2023-11-28
Add end information for SimParticles

AndrewEdmonds11:
It would be useful to have end information for the SimParticles and not just the start info (see Issue #109) so I've added it here. I've kept the names of the start position and momentum branches as ```mom``` and ```pos``` to avoid making a breaking change but I'm open to changing them to ```startmom``` and ```startpos``` for clarity

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,9 @@
 
 TrkAna is a track-based ntuple for Mu2e analyses. Each entry in the ntuple corresponds to a single reconstructed track. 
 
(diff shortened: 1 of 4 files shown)


brownd1978 (approved):
I think it's OK to leave the start momentum as 'mom'.     I see that the TrkAna V4 spreadsheet is out-of-date about the mcsim branch, that should be updated for this and other recent changers

PR #118, opened 2024-01-12
Update count structs

brownd1978:
Note the global hit struct is unfilled, that code got lost in some recent update.  The global track count also is unfilled.


AndrewEdmonds11:
Thanks, Dave. I didn't realize that the hit counts and track counts aren't being filled. I'll track that down and fix it

PR #120, opened 2024-02-06
Added MC truth track parameters

sophiemiddleton:
test with https://github.com/Mu2e/Offline/pull/1182


fcl/prolog_trigger.fcl
```diff
@@ -14,7 +14,7 @@ TrkAnaTrigger : {
     FillCaloMC : false
     RecoCountTag : ""
     FillTriggerInfo : true
-    FillTrkQualInfo : false
+    //FillTrkQualInfo : false
```
AndrewEdmonds11, line 17: You can just remove this line - the parameter no longer exists. I missed deleting it here when doing the recent TrkQual update

AndrewEdmonds11 (approved):
Thanks, Sophie. This looks great. Just one small comment to delete a fcl parameter that no longer exists. As far as I can tell, this doesn't need to wait for the Offline and Production PRs, right?

sophiemiddleton:
I think it needs to wait for the Offline Pr

src/InfoMCStructHelper.cc
```diff
@@ -209,12 +232,13 @@ namespace mu2e {
     siminfo.valid = true;
     if(sp.genParticle().isNonnull())siminfo.gen = sp.genParticle()->generatorId().id();
     siminfo.proc = sp.creationCode();
+    siminfo.stop = sp.stoppingCode();
```
AndrewEdmonds11, line 235: Hi Sophie - I don't see a ```stop``` leaf in the ```SimInfo``` struct in this PR. Did you meain to include an updated SimInfo.hh?

PR #123, opened 2024-02-22
Some changes to sim info and maxr fix

sophiemiddleton:
Includes two features:

* Move proc --> startCode
* maxr fix


sophiemiddleton:
lots of. diffs, I guess I need to rebase to the dev branch, its not an immediate need anyway

AndrewEdmonds11:
Thanks, Sophie. I think it's just because I haven't merged main into v05-dev yet. This PR might look cleaner once I do that. By the way, does the maxr fix not need to go into v4 too?

sophiemiddleton:
yes, it should go into that too I guess, I'm not sure if anyone but me is using it but I could put in two PR's if you'd prefer?

AndrewEdmonds11:
Thanks, Sophie. Two PRs would be great :)

PR #139, opened 2024-04-21
Update to use KalSeedPtrs and add example of using a merged track collection

AndrewEdmonds11:
This PR updates TrkAna to use a collection of KalSeedPtrs rather than a collection of KalSeeds. This makes it easy to combine track collections into one and have a single branch for all tracks. I have added an example fcl configuration file (```TrkAnaReco_mergedKalSeeds.fcl```) that merges all KalSeeds into one collection and puts it all in one ```trk``` branch.

The behavior of TrkAnaReco.fcl hasn't changed. In order to support both one branch and multiple branches, I have added ```MergeKalSeeds``` modules that "merge" a single KalSeedCollection to get a KalSeedPtrCollection. I have validated TrkAnaReco.fcl with the validation script before and after making changes and the histograms are identical as reported by ```valCompare```

### DIFF
--- a/fcl/TrkAnaReco.fcl
+++ b/fcl/TrkAnaReco.fcl
@@ -18,7 +18,7 @@ physics :
   analyzers : { @table::TrkAnaReco.analyzers }
 }
 
-physics.TrkAnaTrigPath : [ PBIWeight, TrkQualDeM ]
+physics.TrkAnaTrigPath : [ @sequence::MergeKKProducersPath, PBIWeight, TrkQualDeM ]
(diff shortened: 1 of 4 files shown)


brownd1978 (approved):
This looks good.  I like that you kept the previous default behavior (separate branches for DeM, ...).   We are considering changes to Offline to combine some reco sequences, which means separate DeM, DeP etc won't be available anymore.  I suggest you should advertise the merged option, as that won't be affected by that change.
Eventually, if the merged option becomes prefered, I would suggest reversing the script dependence, so that TrkAnaReco assumes merged inputs, and TrkAnaRecoUnmerged  would override that with separate inputs.

PR #142, opened 2024-05-07
Added calo clusters information and calo cluster hits informations

pfedeli:
I added the branch calorecoinfo with the general clusters and hits information.


AndrewEdmonds11:
I'm going to close this PR and reopen #141 so that all comments are in one place. (I think that will work because it looks like the most recent commit here was added over there...)

PR #150, opened 2024-06-06
Update script

brownd1978:
This fix adds the necessary indirection.  There are other scripts in TrkAna/fcl that need updating I'll try to get to those as well.


brownd1978:
This PR now depends on Offline PR 1282

AndrewEdmonds11 (approved):

PR #177, opened 2024-06-28
Added python util

sophiemiddleton:
Contains a few well used functions in an interface to allow users to import into a given python script and use these functions.

We intend to evolve it as we develop more examples.


utils/test.py
```diff
@@ -0,0 +1,11 @@
+from mu2epyutil import EvtNtuple_util as evn
+
+def main():
+  test_evn = evn("/exp/mu2e/app/users/sophie/ProductionEnsembles_v2/py-ana/trkana/pass0a.root", "TrkAna", "trkana")
+  vector_test = test_evn.GetVectorXYZ("demfit", "mom", 0)
+  print(vector_test.mag)
+  test_evn.PlotMagValueHist("demfit", "mom", 0, 95, 115, "reco mom at Trk Ent [MeV/c]","log")
+  test_evn.PlotValueHist("demfit", "time", 0, 450, 1700, "fit time at Trk Ent [ns]", 'linear')
```
AndrewEdmonds11, line 8: Could you define a variable for 0s? Just to make it clear that this is the surface ID for the tracker entrance.

In the near future, we should use ```SurfaceId``` directly (via ```pywrap```)

utils/mu2epyutil.py
```diff
@@ -0,0 +1,115 @@
...
+    }, with_name="Vector3D")
+    
+    return trkvect3D
+  
+  def PlotValueHist(self, leafname, vectorreq, sid, low, hi, xaxis_label, scale='linear'):
+    """ make basic plot of magnitude of leafname value at tracker ent, mid or ext (specify sid) """
+    
+    # import code and extract branch
+    tree = self.ImportTree()
+    branch = self.ImportBranches(tree, [leafname])
```
AndrewEdmonds11, line 68: I think we don't want to do this in the plotting functions? Unless ```ImportTree``` and ```ImportBranches``` are smart enough to cache contents

AndrewEdmonds11 (commented):
Thanks for this, Sophie. This looks great! Could you also add a page to the tutorial to point people to this?

Also at some point in the future, this will need to handle iterating through large (or many) files. Is it clear how to do that?

utils/mu2epyutil.py
sophiemiddleton: I mean this is a class so it works fine as far as I can tell, what is your worry?

utils/test.py
sophiemiddleton: Can do

utils/mu2epyutil.py
AndrewEdmonds11: Just that we could end up with the same file open or read multiple times. We can just keep an eye on it

sophiemiddleton:
Hi andy,

I am upgrading the tutorials for the other issue, should I include this in that or here. Yes the many files thing should be doable. I can upgrade it if you want, now or later?

AndrewEdmonds11:
I think for this PR you can just add a one-line page saying something like "We have a python utility package here: link" and link it from the main page or something. Then your upgraded tutorial can be a separate PR

If it's easy to do the many files thing now, then go for it; otherwise, we can just open an issue and come back to it later

sophiemiddleton:
I've added a simple tutorial. Let's open a new issue for the multiple files, I need to look over some examples

sophiemiddleton:
continued on https://github.com/Mu2e/TrkAna/issues/178

PR #181, opened 2024-07-09
Make dictionaries available to ROOT

AndrewEdmonds11:


rlcee:
Shall I add these lines to the last cvmfs install of TrkAna?  Should I also go back a few versions?

AndrewEdmonds11:
Note that users will also need to make sure they are linked to a version of Offline with the same change. If not, you may have issues with the ```MCRelationship``` leaves of ```demmcsim```. For example, I found that they could no longer be scanned on the ROOT command line (```trkana->Scan("demmcsim.prirel._rel")```). I didn't check other ways of accessing

If you don't have a correct version of Offline, these errors wlil appear when opening the ROOT file:
```
Error in <TProtoClass::FindDataMember>: data member with index 0 is not found in class mu2e::MCRelationship
Error in <CreateRealData>: Cannot find data member # 0 of class mu2e::MCRelationship for parent mu2e::SimInfo!
Error in <TProtoClass::FindDataMember>: data member with index 1 is not found in class mu2e::MCRelationship
Error in <CreateRealData>: Cannot find data member # 1 of class mu2e::MCRelationship for parent mu2e::SimInfo!
Error in <TProtoClass::FindDataMember>: data member with index 0 is not found in class mu2e::MCRelationship
Error in <CreateRealData>: Cannot find data member # 0 of class mu2e::MCRelationship for parent mu2e::SimInfo!
Error in <TProtoClass::FindDataMember>: data member with index 1 is not found in class mu2e::MCRelationship
Error in <CreateRealData>: Cannot find data member # 1 of class mu2e::MCRelationship for parent mu2e::SimInfo!
```

I will open an issue to see if there's a way we can get around this

AndrewEdmonds11:
> Shall I add these lines to the last cvmfs install of TrkAna? Should I also go back a few versions?

Thanks, Ray. That would be great. I think as well it would be good to include them in the Offline installs too (see my comment above about MCRelationship)

rlcee:
I made this change to TrkAna v05_03 and v05_04, and Offline v10_30 and v10_31. visible within the hour.  let me know if you need more

PR #193, opened 2024-07-23
Add old track fit parameters to LoopHelixInfo as a stopgap

AndrewEdmonds11:
This PR adds back the old track parameters to the ```demlh``` branch: ```d0``` and ```tanDip```.

Note that the ```d0``` added here is unsigned. I don't think this is an issue because the most recent sensitivity update used the magnitude of the signed ```d0``` for a cut.

Also note that these are just a stopgap until we have proper physical variables that can be used in the analysis. I will add an issue so that we remove them at some point in the future

### DIFF
--- a/inc/TrkInfo.hh
+++ b/inc/TrkInfo.hh
@@ -36,9 +36,11 @@ namespace mu2e
   };
 
   struct LoopHelixInfo {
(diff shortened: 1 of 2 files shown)


brownd1978 (approved):
looks good

PR #197, opened 2024-08-06
Create SurfaceStepInfos from SurfaceSteps

brownd1978:
SurfaceStepInfos are added as a new branche. In principle these replace all existing branches made from StepPointMCs.
This PR also adds the sequence needed to build SurfaceSteps from MDC2020 output. In the next sim production that producer will be run as part of production.
Note this PR depends on Offline PR 1321.


AndrewEdmonds11 (commented):
Thanks, Dave. This looks good. Just one request: could you add an example Draw command to the tutorial page? It doesn't need to be a full exercise (or even on it's own page) but just so we have something we can incorporate later

PR #205, opened 2024-09-23
Trkqual changes

sophiemiddleton:
- restructures the trkqual branch such that it is similar to the dem branch.
- more to follow for other similar branches


sophiemiddleton:
solves https://github.com/Mu2e/TrkAna/issues/146

PR #232, opened 2024-11-21
Fix build failure in nightly

AndrewEdmonds11:
Related to change in CrvDigi in Mu2e/Offline#1368


ehrlich-uva (approved):
This change became necessary, because we moved from a fixed-length std::array of samples to a vector to accommodate digis with different numbers of samples.

sam-grant (approved):

PR #236, opened 2024-12-03
Pyutils: first round of pyutils for v6

sophiemiddleton:
A set of classes and examples to help python based analysis.

not final, will refine in second PR


utils/pyutils/README.md
```diff
@@ -0,0 +1,197 @@
...
+
+The Analysis python Environment is maintained by the L4 for Analysis Environment (currently Sam Grant).
+
+To activate the environment:
+
+```
+source /cvmfs/mu2e.opensciencegrid.org/env/ana/current/bin/activate
+```
+
+If a package you require is not present, please contact Sam Grant (L4 for Analysis Env).
```
AndrewEdmonds11, line 17: I think we should have people post requests to the Slack channel

example-analysis-scripts/pyutils-examples/example_cuts.py
```diff
@@ -0,0 +1,86 @@
+#! /usr/bin/env python
```
AndrewEdmonds11, line 1: I've been putting rooutil examples under ```rooutil/```. I think I'd like to get rid of the top-level ```example-analysis-scripts``` directory but perhaps it's good to have something obvious at the top-level for people to see...

AndrewEdmonds11 (approved):
Thanks, Sophie! A couple of in-line comments. Let me know if you want to address these now or in the next PR. One other thing to think about is whether there is some validation script that could be written to make sure we don't break pyutils as we develop it

example-analysis-scripts/pyutils-examples/example_cuts.py
sophiemiddleton: good point, I think examples makes most sense in the example-analysis-script and keep the utils as their own thing but upto you. Let me know and I will move

AndrewEdmonds11: Yeh, I think that's right. Do we also want to move the invidual utils to the top-level? So we would have something like:
```
EventNtuple/
| examples/
| inc/
| src/
| pyutil/
| rooutil/
| validation/
```

sophiemiddleton: I think that could work, they are separate entities

AndrewEdmonds11: Sounds good. Let's do that in the next PRs. I'll merge this now

Issue #235, opened 2024-11-25
Time order the elements in trksegs and trksegsmc


AndrewEdmonds11:
As suggested by @lborrel, it would be good if the segments in ```trksegs``` and ```trksegsmc``` are time ordered. At the moment, there doesn't seem to be an order.

AndrewEdmonds11:
Done in #244

PR #244, opened 2025-01-23
Timeorder of trksegs

sophiemiddleton:
Time ordering of TrkSegs

Time ordering of TrkSegMCs

to solve https://github.com/Mu2e/EventNtuple/issues/235


sophiemiddleton:
Tested with print statement, ordering works. Will commit the MC changes too

PR #255, opened 2025-02-11
Updates for calibration

brownd1978:
The new track lists put dem and dep together, so the calib output also has that. We could add filtering to select just dem, but for now I'm assuming that will be done in the analysis (cosmic events contain roughly even numbers of dem and dep fits (most dep are really muons).


bonventre (approved):

AndrewEdmonds11 (approved):
Thanks, Dave. At some point we might want to extend the "write events only with tracks" to other detectors but that's beyond the scope of this PR

PR #260, opened 2025-03-04
start cmake build

rlcee:


src/CMakeLists.txt
```diff
@@ -0,0 +1,72 @@
...
+      Offline::KinKalGeom
+      Offline::Mu2eKinKal
+      Offline::TrackerConditions
+      Offline::Mu2eUtilities
+      Offline::GeneralUtilities
+
+    LIBRARY_NAME eventntuple
+)
+
+art_dictionary( NO_CHECK_CLASS_VERSION # For some reason this segfaults
```
AndrewEdmonds11, line 32: Can you clarify this comment? The code segfaults without this ```NO_CHECK_CLASS_VERSION``` or something else?

AndrewEdmonds11 (approved):
Thanks for adding this, Ray. I just have one question below

src/CMakeLists.txt
rlcee: Actually I can't! that comment is from cut-n-paste from other cmake files.  We might find the source or post the question, if you think it is important.  Or I can remove the comment.

AndrewEdmonds11: No worries. I'm sure it's fine. We can always work it out later if we need to

PR #262, opened 2025-03-31
Branch documentation update

lborrel:
Update to the branch documentation to reflect what the structure of EventNtuple looks like.
There are a couple of missing branches that I don't fully understand the structure:
- trkcalohitmc has 2 vectors per event, I don't know what they are
- trkhits and trkhitsmc cannot be accessed through python so I didn't double check the structure
- I didn't find trkmcsci and trkmcssi in my sample


doc/branches.md
```diff
@@ -1,39 +1,82 @@
...
+
+| branch | structure | explanation | leaf information |
+|--------|-----------|-------------|------------------|
+| evtinfo |  Single objects branch |   event-level information| [see EventInfo.hh](../inc/EventInfo.hh)
+| evtinfomc |  Single objects branch |   MC-truth event-level information| [see EventInfoMC.hh](../inc/EventInfoMC.hh)
+| hitcount |  Single objects branch |   counts of different hit types in an event| [see HitCount.hh](../inc/HitCount.hh)
+| tcnt |  Single objects branch |   counts track types and track-related quantities in an event (MARKED FOR REMOVAL)| [see TrkCount.hh](../inc/TrkCount.hh)
+## Track Branches
+
+These branches contain 4 elements per event corresponding to different Kalman fit hypotheses:
```
AndrewEdmonds11, line 15: Technically, it could be less than 4 elements if e.g. the upstream fits completely fail or are not done. It could also be more than 4 elements if there are e.g. two real downstream tracks to fit. I think we want to make the point that there will be multiple tracks in each event and they could be the same track just with different fit hypotheses...

lborrel:
I implemented the minor fixes you suggested, you can merge it and we can update the documentation once the new branches are in

PR #265, opened 2025-04-07
Updates to pyimport and pyprint, addition of pyread and pyprocess

sam-grant:
# `pyutils` updates 

## Changes to `pyimport`, addition of `pyread` and `pyprocess`

Added two new `pyutils` modules, `pyread` and `pyprocess`, and modified the existing `pyimport`.  The primary purpose of these changes is to allow users to easily and efficiently import EventNtuple files and datasets into a array which is structured according to need. These additions are described in detail in [DocDB-52297](https://mu2e-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=52297). 

>**Note**: Following review, I made changes to the README and added three tutorial notebooks. 

### A point raised in the developed meeting

For large datasets, concatenating data across hundreds or thousands of files will run into memory issues. The solution to this is to accumulate results in user defined histograms rather than storing the full array for each file. I will add this to functionality to the `Import` class in a future PR.

###  Changes to `pyprint`

I have added two new optional arguments to the class init: `verbose` (bool) and `precision` (int). When `verbose` is `True`, it will print out the full value arrays for each branch without the automatic truncation from awkward, where each element has a number of decimal points specified by `precision`. 

**Default behaviour:** 

```python
printer = Print()
printer.print_n_events(data, n_events=1)
```

```
[pyprint] ---> Printing 1 event(s)...

-------------------------------------------------------------------------------------
event: 1
trksegs.mom.fCoordinates.fX: [[-17.5, 3.86, -32.1, -2.34, -34.4, -52.5, -20.6], ..., [-10.3, -77, ..., 58]]
-------------------------------------------------------------------------------------
```

**With `verbose=True`** 

```python
printer = Print(verbose=True, precision=1)
printer.print_n_events(data, n_events=1)
```

```
-------------------------------------------------------------------------------------
event: 1
trksegs.mom.fCoordinates.fX: [[-17.5, 3.9, -32.1, -2.3, -34.4, -52.5, -20.6], [-51.9, -96.8, -82.4, -100.1, -96.7, -99.5, -94.1, 8.1, 23.9, 77.8, 10.4], [-10.7, -78.1, -18.9, -7.8, 89.8, 70.2, 92.8, 98.7, 100.8, 98.2, 86.8, 96.5, -12.9, 35.7, 67.8], [16.3, -7.0, -31.8, -14.7, -17.0, -35.9, -16.3], [92.5, -99.6, -88.7, -104.3, -87.1, -86.3, -97.1, 7.8, 25.9, 79.3, 10.6], [-10.3, -77.0, -30.4, -7.9, 91.2, 92.2, 97.5, 99.5, 80.7, 96.4, 78.8, 89.6, 93.4, -11.7, 35.5, 58.0]]
-------------------------------------------------------------------------------------
```


sophiemiddleton (approved):

sophiemiddleton:
thanks @sam-grant these look great. Can you also add some details to the README and perhaps an example of how to use them too?

sam-grant:
I have added the "location" and "schema" arguments to `pyimport` and `pyread`. These are needed for generating the correct xroot URL when reading files remotely. Previously I had the location hardcoded to "tape" and the schema to "root". The arguments are passed to the classes during initialisation. I also moved `use_remote` to the initialisation to keep things a bit cleaner.

sam-grant:
And I will add to the README as well.

sophiemiddleton (changes requested):
Needs updates to the examples to make sure people can use these

sam-grant:
Added the first tutorial notebook, updated README, and made some changes to all modules to improve consistency.

sam-grant:
Added third tutorial notebook.

sophiemiddleton (approved):

sam-grant:
There are a few typos in the tutorials and the README, I'll sort those out.

PR #274, opened 2025-05-12
Switch from SelectRecoMC -> SelectReco

brownd1978:
This PR changes the configuration to be compatible with MC-agnostic reco output that will appear in the next round of production. It is a breaking change as I've not found a way to wildcard the tag.


AndrewEdmonds11 (approved):
Thanks, Dave. This looks good. I'll merge it in when we start the new production

PR #275, opened 2025-05-13
Add calo branches

giro94:
Adding calorimeter branches: clusters, hits, recodigis.

Plan is to have separate branches for each struct type.
Clusters will have vector of indexes of hits.
Hits will have vector of indexes of recodigis.
Hits (digis) will have a single index of the cluster (hit) they belong to.


AndrewEdmonds11 (commented):
Thanks for this, Paolo! It all looks OK to me. Would you mind also doing step 8 and the first part of step 11 of [these instructions](https://github.com/Mu2e/EventNtuple/blob/main/doc/developers.md#Adding-a-branch-to-EventNtuple) to add the new branches to the ```ntuplehelper```?

We can discuss who does what in the other steps but that can all be done in a separate PR

sophiemiddleton (approved):
I'm happy with the code just by eye, i'd like to test it just to get a feel for the new branch

PR #278, opened 2025-06-12
Updates for CRV

AndrewEdmonds11:
This PR makes two changes related to the CRV.

### CrvPlaneY = 0 Fix
The leaf ```crvcoincsmcplane.pos.fCoordinates.fY``` was always being set to 0. This is now fixed

### Record MC-truth plane info for all particles
The branch ```crvcoincsmcplane``` only recorded information if the primary was a positive or negative muon.

One situation I met that I don't think is covered by the various ```MCRelationship``` and ```originParticle``` functions is that the primary cosmic-ray neutron produces a "daughter" neutron that crosses the plane. I think we want to treat this "daughter" neutron as the same as the "parent" neutron so that's what I've implemented. If this situation is covered elsewhere, then let me know

Here is a before and after of event 1202, subrun 60134, event 103199, which has a cosmic neutron as a primary
Before:
```
ntuple->Scan("evtinfo.run:evtinfo.subrun:evtinfo.event:crvcoincsmcplane.pos.fCoordinates.fY:crvcoincsmcplane.pdgId", "")
***********************************************************************************
*    Row   * Instance * evtinfo.r * evtinfo.s * evtinfo.e * crvcoincs * crvcoincs *
***********************************************************************************
*        0 *        0 *      1202 *     60134 *    103199 *           *           *
***********************************************************************************
```

After:
```
 ntuple->Scan("evtinfo.run:evtinfo.subrun:evtinfo.event:crvcoincsmcplane.pos.fCoordinates.fY:crvcoincsmcplane.pdgId", "")
***********************************************************************************
*    Row   * Instance * evtinfo.r * evtinfo.s * evtinfo.e * crvcoincs * crvcoincs *
***********************************************************************************
*        0 *        0 *      1202 *     60134 *    103199 *      2653 *      2112 *
***********************************************************************************
```

### Fcl Options for CRV plane heights
Added in new fcl prolog ```prolog_crv_planes.fcl```

### DIFF
--- a/fcl/prolog.fcl
+++ b/fcl/prolog.fcl
@@ -3,6 +3,7 @@
 #include "Offline/CommonMC/fcl/prolog.fcl"
 #include "Offline/fcl/TrkCaloDt.fcl"
 
+#include "EventNtuple/fcl/prolog_crv_planes.fcl"
 #include "EventNtuple/fcl/prolog_mc_reweighters.fcl"
 
 BEGIN_PROLOG
@@ -237,7 +238,7 @@ EventNtupleMaker : {
   CrvDigiMCsTag : "compressRecoMCs"
   CrvDigisTag : "SelectReco"
   CrvCoincidenceClusterMCAssnsTag : "CrvCoincidenceClusterMCAssns"
-  CrvPlaneY : 2653
+  CrvPlaneY : @local::crvPlaneY.CRV_T
   FillMCInfo : true
   FillTrkPIDInfo : false
   FillHitInfo : true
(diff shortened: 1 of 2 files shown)


oksuzian:
Would it make sense to add fcl with CrvPlaneY set to the height of DS hatch blocks?

AndrewEdmonds11:
Yep, added in commit [0244f62](https://github.com/Mu2e/EventNtuple/pull/278/commits/0244f626f9d3183904524de4fc36e00e32172d39). Let me know if the values need changing

PR #283, opened 2025-07-07
Update from_digi script

brownd1978:
This includes some minor improvements:
- use R() instead of sqrt(mag2()) for XYZVector payload.
-  removes the (non-functional) earliest and latest intersection flg, which can be extracted during analysis anyways.
- add a flag for trkstrawinfomc that are matched to trkstrawinfo. This isn't strictly necessary but can be helpful
-use combined KK reco sequence output in the from_digi script

I note that the rest of the config assumes the KKDeM, KKDeP split outputs, which is inconsistent with the head of Offline and the most recent MDC2020 samples. This might be a good time to reconsider having split trees. Some reworking of TrkQual calculation is also needed, which may impact how it is trained, or at least how the module loads the training.

### DIFF
--- a/fcl/from_dig-DeCalib.fcl
+++ b/fcl/from_dig-DeCalib.fcl
@@ -46,12 +46,12 @@ physics :
 physics.end_paths : [ EndPath ]
 physics.trigger_paths : [ RecoPath ]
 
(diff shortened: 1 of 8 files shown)


brownd1978:
Thanks Andy, I didn't realize there was a default TrkQualDe instance, I updated the script to use that.

PR #308, opened 2025-10-04
Updates to RooUtil

sophiemiddleton:
Several updates to rooutils to make analysis easier:

* New roodisplay module allows for easier call to the display
* Added cutflow module in my last PR, now I've updated it to include a histogram option
* Some updates to common cuts to help analysis of MDS

### DIFF
--- a/rooutil/inc/common_cuts.hh
+++ b/rooutil/inc/common_cuts.hh
@@ -112,7 +112,7 @@ bool has_reco_step(TrackSegment& segment) { // track fit segment has an reco Sur
 
 //+ Track Cuts - Direction
 bool is_downstream(Track& track) { // reconstructed fit segment at tracker middle is going in +z direction
-  auto tracker_middle_segments = track.GetSegments([](TrackSegment& segment){ return tracker_middle(segment) && !has_mc_step(segment) && has_reco_step(segment); });
+  auto tracker_middle_segments = track.GetSegments([](TrackSegment& segment){ return tracker_middle(segment) && has_reco_step(segment); });
   if (tracker_middle_segments.size() > 1) {
     std::cout << "WARNING: " << tracker_middle_segments.size() << " track segments at tracker middle. Something might have gone wrong..." << std::endl;
     return false;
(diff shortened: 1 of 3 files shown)


rooutil/inc/common_cuts.hh
```diff
@@ -191,8 +191,8 @@ bool three_of_four_coinc(CrvCoinc& crv_coinc) { // CRV coincidence has exactly t
 }
 
 //+ Combined Track & CrvCoinc Cuts
-bool track_crv_coincidence(TrackSegment& segment, CrvCoinc& crv_coinc) { // time difference between track segment and crv_coinc is less than 250 ns
-  if ( segment.trkseg->time - crv_coinc.reco->time > 200 && segment.trkseg->time - crv_coinc.reco->time < 450) { return true; }
+bool track_crv_coincidence(TrackSegment& segment, CrvCoinc& crv_coinc) { 
+  if ( abs(segment.trkseg->time - crv_coinc.reco->time) > 150 ) { return true; }
```
AndrewEdmonds11, line 195: Is the sign the right way around here? This is saying if the time difference is more than 150 ns, then it's true that there's a tracker-CRV coincidence. Maybe there's another way of reading the function name so we might want to change it

rooutil/inc/cutflow.hh
```diff
@@ -18,14 +18,23 @@ usage:
...
+
+#include "TH1.h"
+
 #include <format>
 // Structure to hold cutflow data
 struct CutflowEntry {
+    /*
+    Flexible struct. Two field "cut_name" e.g t>600 ns and "passed_count" e.g. 1000 events
+    */
     std::string cut_name;
```
AndrewEdmonds11, line 30: No need to implement in this PR but could we use the cut functions here? Than the user doesn't need to pass in the passed_count and we can get this function to do the counting too...

utils/pyutils
AndrewEdmonds11, line 1: What's the reason for adding pyutils as a submodule?

AndrewEdmonds11 (commented):
Thanks for all this, Sophie! I've made a few comments below. Could you also add some words to the README and an example macro that describes and uses the cutflow and roodisplay?

rooutil/inc/common_cuts.hh
sophiemiddleton: yes, I think its just a difference in how I was using it. I will reverse the sign for it to make more sense

sophiemiddleton:
@AndrewEdmonds11 I will target the example in the next round

PR #312, opened 2025-11-07
trigger branch

sophiemiddleton:
Introduces new trigger branch.

- uses trigger navigator to make the branches (one per trigger) 
- fills at event level
Screenshot 2025-11-07 at 13 43 52

### DIFF
--- a/src/InfoStructHelper.cc
+++ b/src/InfoStructHelper.cc
@@ -528,7 +528,7 @@ namespace mu2e {
     digiinfo.caloRecoDigiIdx_ = recodigiIdx;
     digiinfos.push_back(digiinfo);
   }
-  
-  
+
+
 
 }


src/EventNtupleMaker_module.cc
```diff
@@ -129,8 +130,8 @@ namespace mu2e {
         using Comment=fhicl::Comment;
 
         // General control and config
-        fhicl::Atom<int> diag{Name("diagLevel"),1};
-        fhicl::Atom<int> debug{Name("debugLevel"),0};
+        fhicl::Atom<int> diag{Name("diagLevel"),0};
+        fhicl::Atom<int> debug{Name("debugLevel"),5};
```
AndrewEdmonds11: Could you change the default debug level back to 0?

src/EventNtupleMaker_module.cc
```diff
@@ -325,9 +328,11 @@ namespace mu2e {
       FType _ftype = Unknown;
       std::vector<std::string> fitNames = {"Unknown", "LoopHelix","CentralHelix","KinematicLine"};
 
+      // for trigger branch:
+      bool firstEvent = true;
       // helper functions
       void fillEventInfo(const art::Event& event);
-      void fillTriggerBits(const art::Event& event,std::string const& process);
+      void fillTriggerBranch(const art::Event& event,std::string const& process, bool firstEvent, TTree* &_ntuple);
```
AndrewEdmonds11: I'm surprised you need to pass the ntuple by argument. It should be accessible from within fillTriggerBranch alreadysince it is a member variable of the class. Is there something I'm missing?

utils/pyutils
AndrewEdmonds11: Please remove

AndrewEdmonds11 (commented):
Thanks for putting this in, Sophie. I have a few inline comments below

src/EventNtupleMaker_module.cc
```diff
@@ -924,40 +926,27 @@ namespace mu2e {
...
-      if(_conf.debug() > 1){
-        TriggerResultsNavigator tnav(trigResults);
-        tnav.print();
-      }
+
+    for (unsigned int i = 0; i < tnav.getTrigPaths().size(); ++i) {
+        const std::string path = tnav.getTrigPathName(i);
+        bool accepted = tnav.accepted(path);
+        _triggerResults.triggerResults[i] = accepted;
+        _triggerArray[i] = static_cast<UChar_t>(_triggerResults.triggerResults[i]);
```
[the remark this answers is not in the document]
sophiemiddleton: there was some issue with the memory address, this was the hacky work around I told you about.

We could perhaps delete the triginfo class and just use the triggerarray, but the class has a bit more of a use-case

src/EventNtupleMaker_module.cc
sophiemiddleton: there was a reason previously but now I see its not needed.

PR #323, opened 2025-12-09
Add Trigger rooutil object 

michaelmackenzie:
Add a `rooutil::Trigger` object to hold the trigger information and include some simple trigger functions.


michaelmackenzie:
I added a function in the Trigger object to test a pattern against trigger names. With this, one could do `trigger.FiredByTag("TrkDe_80")` to check the 80 MeV/c triggers for any pattern recognition algorithm or D0 selections, for example.

PR #360, opened 2026-03-25
Update location of data files

brownd1978:
This is related to issue 359, that EventNtuple depends on TrkDiag. As TrkDiag was mirgated to ArtAnalysis the immediate fix is to update the file path, but a longer term solution is still needed.


AndrewEdmonds11 (approved):
Thanks, Dave!

PR #376, opened 2026-06-11
CRV pulse hit index to replace pulses vector

RobMina:
Replace the std::vector<CrvPulseInfoReco> in CrvHitInfoReco with a flat int crvHitIndex in each CrvPulseInfoReco struct. The std::vector was not straightforward to unpack in pyutils and resulted in redundant data (the same pulse data stored in two separate branches). The new int index allows the pulses collected into a hit to be associated to it for analysis. A new KeepUnclusteredPulses FHiCL parameter (default false) allows to filter out reco pulses that were not collected into a hit.

In validation, discovered a segmentation fault and out-of-bounds vector element accesses. Those have been guarded so that test_fcls.sh runs cleanly.


validation/ceSimReco.fcl
```diff
@@ -7,3 +7,5 @@
 
 physics.producers.KKDe.KKFitSettings.SaveHitCalibInfo : true
 outputs.Output.fileName : "mcs.owner.val-ceSimRecoVal.dsconf.seq.art"
+// Production/Validation/ceSimReco drops CaloDigis; append keep so it wins over drop
+outputs.Output.outputCommands[50]: "keep mu2e::CaloDigis_CaloDigiMaker_*_*"
```
AndrewEdmonds11: If the intent is to append, then I suggest doing:
```
outputs.Output.outputCommands : [ @sequence::outputs.Output.outputCommands, "keep mu2e::CaloDigis_CaloDigiMaker_*_*" ]
```
Just in case the output commands in ceSimReco.fcl itself changes in future

rooutil/inc/Event.hh
```diff
@@ -129,6 +129,8 @@ namespace rooutil {
       tracks.clear();
       for (int i_track = 0; i_track < nTracks(); ++i_track) {
         if (debug) { std::cout << "Event::Update(): Creating Track " << i_track << "... " << std::endl; }
+        if (trksegs == nullptr || i_track >= static_cast<int>(trksegs->size())) continue;
```
AndrewEdmonds11: This was causing the segfault you mentioned? In which case we have a deeper issue because there we deliberately construct the ```trk```, ```trksegs```, and ```trkcalohit``` branches to have the same length

rooutil/inc/Event.hh
```diff
@@ -154,7 +156,7 @@ namespace rooutil {
       crv_coincs.clear();
       for (int i_crv_coinc = 0; i_crv_coinc < nCrvCoincs(); ++i_crv_coinc) {
         CrvCoinc crv_coinc(&(crvcoincs->at(i_crv_coinc)));
-        if (crvcoincsmc != nullptr) {
+        if (crvcoincsmc != nullptr && i_crv_coinc < static_cast<int>(crvcoincsmc->size())) {
```
AndrewEdmonds11: Are ```crvcoincs``` and ```crvcoincsmc``` branches not lock-step arrays?

rooutil/inc/Event.hh
```diff
@@ -218,7 +221,7 @@ namespace rooutil {
     }
 
     template <typename T> void UpdateObject(T*& object, std::vector<T>* object_ptr, int index, bool debug = false) {
-      if (object_ptr != nullptr) {
+      if (object_ptr != nullptr && index < static_cast<int>(object_ptr->size())) {
```
AndrewEdmonds11: I think we might want to print a warning that the index is too high for the vector rather than just skip it. It would mean that something has gone out of sync

rooutil/inc/Track.hh
```diff
@@ -82,19 +83,22 @@ namespace rooutil {
   
       if (trksegpars_lh != nullptr) { // if we LoopHelix info
         if (debug) { std::cout << "Track::Update(): Updating trksegpars_lh..." << std::endl; }
-        for (int i_segment = 0; i_segment < nSegments(); ++i_segment) {
+        const int n_lh = std::min(nSegments(), static_cast<int>(trksegpars_lh->size()));
```
AndrewEdmonds11: I don't understand why we need this change? Is the ```nSegments()``` function not correct?

src/EventNtupleMaker_module.cc
```diff
@@ -986,6 +989,7 @@ namespace mu2e {
         //Find and link the existing hits
         if (_fillcalohits){
           for (const auto& hit : cluster.caloHitsPtrVector()){
+            if (!hit) continue;
```
AndrewEdmonds11: I don't understand this change

src/InfoStructHelper.cc
```diff
@@ -424,6 +424,10 @@ namespace mu2e {
     if (kseed.hasCaloCluster()) {
       auto const& tch = kseed.caloHit();
       auto const& cc = tch.caloCluster();
+      if (!cc) {
```
AndrewEdmonds11: This case should be being covered at the end of the function. Are you seeing a problem with the caloCluster not existing even though ```kseed.hasCaloCluster()``` is true?

AndrewEdmonds11 (commented):
Thanks for the PR, Rob! I have a few in-line comments below. There are a couple of things where the underlying cause might be elsewhere in the code so we should work out whether these are real fixes or just sticking plasters

rooutil/inc/Event.hh
ehrlich-uva: In Offline, CrvCoincidenceClusters and CrvCoincidenceClustersMC are in lock-step (if MC exists). However, in EventNtuple/src/CrvInfoHelper.cc line 114, we have a check that only results in a cout statement if a mismatch is found. I think it should throw an exception there, and perhaps here as well.

ehrlich-uva (approved):
I only looked at the CRV part of this PR.
I think this is a better way of dealing with the CrvRecoPulses.

RobMina:
@AndrewEdmonds11 most of the guards were unnecessary. That was an overzealous LLM agent papering over the real problem, which is an incorrect assumption about the calohits and calohitsmc branches being aligned. I preserved the original behavior of matching each reco calo hit to the MC calo hit with the same index while preventing an out of bounds vector access, but the real problem is upstream where the association object between the compressRecoMCs and the CaloHitMaker hits is never produced or stored, so there's no handle in the art file for EventNtuple to properly match MC truth hits with reco hits.

AndrewEdmonds11:
@RobMina I think there is just one of Ralf's comments to address and then I'm afraid I merged some big changes for the fhicl configuration updates in #375 that has produced some merge conflicts. Let me know if you need a hand resolving those

RobMina:
Thanks @AndrewEdmonds11 ! I resolved Ralf's comment and merged in the updates from main.

PR #378, opened 2026-06-17
CRV multiple truth planes

RobMina:
Enable multiple configurable CRV truth planes. Default behavior is unchanged. Give options for KPP/run2 geometry studies and add a new parameter file for processing MC truth branches from KPP cosmic datasets.


src/CrvInfoHelper.cc
```diff
@@ -200,22 +202,29 @@ namespace mu2e
...
-              MCInfoPlane.emplace_back(trajectorySimParticle->pdgId(),
-                  trajectoryPrimaryParticle->pdgId(),
-                  trajectoryPrimaryParticle->startMomentum().e(),
-                  tdet->toDetector(trajectoryPrimaryParticle->startPosition()),
-                  tdet->toDetector(planePos),
-                  planeDir,
-                  planeTime,
-                  planeKineticEnergy,
-                  0);  //unused
+              int axis = (k<crvPlaneAxes.size()) ? crvPlaneAxes[k] : 1;  // 0=x (L/R sides), 1=y (top), 2=z; default y
```
AndrewEdmonds11: I would rather not have defaults hidden in the code. They should be given explictly in the fhicl

src/EventNtupleMaker_module.cc
```diff
@@ -237,7 +237,9 @@ namespace mu2e {
         fhicl::Atom<bool>          keepUnclusteredPulses {Name("keepUnclusteredPulses"), Comment("If false, crvpulses stores only pulses assigned to CRV coincidence clusters"), false};
         fhicl::Atom<bool>          fillDigis             {Name("fillDigis"),             Comment("Fill CRV digi branch")};
         fhicl::Atom<art::InputTag> digisTag       {Name("digisTag"),        Comment("Tag for CrvDigi collection")};
-        fhicl::Atom<double>        planeY         {Name("planeY"),          Comment("Y of center of top layer of CRV-T counters (mm)")};
+        fhicl::Sequence<double>      planeCoords{Name("planeCoords"),Comment("Coordinate (mm, Mu2e frame) of each CRV truth plane; the axis is chosen by planeAxes")};
+        fhicl::Sequence<int>         planeAxes  {Name("planeAxes"),  Comment("Axis each truth plane is perpendicular to: 0=x (L/R sides), 1=y (top), 2=z")};
```
AndrewEdmonds11: I'm not a huge fan of this sort of thing where you need to know what bare integers correspond to. If you can think of a quick solution, then implement it but it's not critical.

fcl/prolog_crv_planes.fcl
```diff
@@ -1,9 +1,26 @@
...
-
-  CRV_T : 2653
+  CRV_T   : 2653
   DSHatch : 4287
 }
 
+# Run 2 configuration: Side-face x-values (Mu2e coords, mm) for vertical CRV truth planes (constant x).
+# From crv_counters_v10.txt firstCounter x: R1/R2 = -6441.2, L1/L2 = -1366.8.
+crvPlaneX : {
+  CRV_R   : -6441.2
```
AndrewEdmonds11: Ideally we would have the code get these from the GeometryService rather than define them in fhicl. That's beyond the scope of this PR but I will open an issue for it

AndrewEdmonds11 (commented):
Thanks, Rob. I have a few comments below but only one is crticial before I merge.

RobMina:
Thanks @AndrewEdmonds11 ! New implementation just uses a list of string sector names in the FCL and pulls orientation/position from GeometryService at run time.

AndrewEdmonds11 (approved):
Great! Thanks, Rob

PR #387, opened 2026-07-27
Split vector<MCRelationship> branches for uproot

AndrewEdmonds11:
This PR splits leaves that are ```vector<MCRelationship>``` since uproot cannot read the branches that contain them

### DIFF
--- a/inc/CaloClusterInfoMC.hh
+++ b/inc/CaloClusterInfoMC.hh
@@ -14,7 +14,8 @@ namespace mu2e
     float eprimary; // primary particle true energy in this cluster
     float tprimary; // primary particle average time
     std::vector<int> simParticleIds; // list of simparticle ids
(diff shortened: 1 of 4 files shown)


sophiemiddleton (approved):
Looks good. Should we do something about this too:

https://github.com/Mu2e/EventNtuple/blob/main/inc/SimInfo.hh#L30

More generally MCRelationship doesnt interface well with python. We could make a new issue for this if necessary

AndrewEdmonds11:
So the crash we were seeing was because of ```vector<MCRelationship>``` leaves. I don't see the same crash when I try to read the ```trkmcsim``` branch. Do you have a code snippet that shows the issue you see?

sophiemiddleton:
For me there is no crash in this case. I'm just thinking in terms of consistency. It is not an urgent issue

PR #392, opened 2026-07-31
Flatten calo-entrant truth into calohitsmc.entrantSimIds

zwl0331:
Companion to Mu2e/Offline#1911, and requires it (this uses the new CaloHitEntrant product), so I am leaving this as a draft until that one merges. The two together replace #366.

This adds an `entrantSimIds` branch to `calohitsmc`: per hit, the calo-entrant SimParticle ID for each contribution, aligned one-to-one with `simParticleIds`. A value of -1 means the producer had no entrant for that deposit.

The fill is controlled by a new `calo.mc.entrantTag` config which defaults to empty, meaning off. Existing configs are untouched and produce identical output. When the tag is set the fill is strict: a missing product, a wrong collection size, a per-hit count mismatch, or an entrant entry that does not reference the exact CaloHitMC being flattened (checked through the product's `Ptr<CaloHitMC>` back-reference, ProductID and index) is an exception, not a warning, so you cannot silently end up with half-filled or wrong-collection truth.

One scope note: `entrantSimIds` is aligned within `calohitsmc`. Matching `calohits` entries to `calohitsmc` entries is a separate, pre-existing question (`caloHitIdx_` / `crystalID_`, cf. 125a71a); this branch does not depend on that join and does not change it. If the `calohitsmc` ordering or the join ever changes, the new column just follows along.

Validation: 200 MDC2025 events (run 001430) with the producer plus this fill on current main. The flattened entrant IDs agree with my published Python pipeline for all 6,554 contributions (3,903/3,903 MC hits matched across ntuple formats). I also checked the April-format version of this change, where the downstream GNN graph labels come out byte-identical (341/341 disk-graphs); on current main I validated at the entrant-root level, since the reco/MC hit join is the separate question above.

### DIFF
--- a/inc/CaloHitInfoMC.hh
+++ b/inc/CaloHitInfoMC.hh
@@ -21,6 +21,7 @@ namespace mu2e
     std::vector<float> momentumIns; // list of the momentum of the SimParticle when entering in the disk
     std::vector<int> simParticleIds; // list of simparticle ids
     std::vector<MCRelationship> simRels; // relationship to the particle that deposited the most energy in the calo Hit
+    std::vector<int> entrantSimIds; // calo-entrant (shower originator) SimParticle id per deposit, aligned with simParticleIds; filled only when calo.mc.entrantTag is configured
     int clusterIdx_; // Cluster index
     int caloHitIdx_; // index into calohits branch, -1 if unset; calohitsmc is NOT index-aligned with calohits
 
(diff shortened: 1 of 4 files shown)


giro94 (approved):
Great idea! Thanks

sophiemiddleton:
@zwl0331 what is the status of this work?

